Issue #6, opened 2017-10-10
Specifying the computational environment


dhimmel:
@publicus you may want to create a small PR to specify the computational environment. You can also specify it with your first code-containing PR. The idea is that we want to make sure the code runs today and in a year from now. And specifying the version of all your software is important towards those ends.

I'd recommend conda, like [we're using](https://github.com/greenelab/scihub/blob/master/environment.yml) for `greenelab/scihub`. Conda environments support both Python and R. They can be a pain in the ass and sometimes break, so don't be shy.

jglev:
Thanks, @dhimmel! Conda took a bit of getting used to, but I think I've got the hang of it now. It seems similar overall to `virtualenv`, which I'm used to using. I've used `conda env export > environment.yml` following the [Conda documentation](https://conda.io/docs/user-guide/tasks/manage-environments.html#sharing-an-environment) to create a YAML list of packages, and will include it in my next in-progress PR.

I've been using another project with a lot of conceptual overlap with this project to get going again with Python, and am diving fully into this now, having re-familiarized myself with the language and workflow.

dhimmel:
> I've used `conda env export > environment.yml` following the Conda documentation to create a YAML list of packages, and will include it in my next in-progress PR.

Cool, note that environment specs created using `conda env export` will generally only work on the OS which they're created. One main issue is that they include build numbers. In addition, they include many dependent packages. One thing I do to improve OS interoperability is to remove build numbers and only specify packages that I explicitly import or use.

jglev:
That's good to know, thanks! Is there a way to do that from within Conda, or do you write something novel to strip build numbers (and the rest)?

On that note, taking, for example, `sqlalchemy=1.1.9=py36_0`, is the build number you're referring to `py36_0`?

dhimmel:
The format is `channel::package=version=build`. I'd omit `=build` but specify `channel`, which will usually be `anaconda` or `conda-forge`. By default, the packages will all be from the `anaconda` channel.

> is the build number you're referring to py36_0

yep

> Is there a way to do that from within Conda

No, it's an area where conda doesn't document the issue well and has poor solutions. Usually, I just build my environment.yml file manually, one package at a time, rather than working backwards from an export.

jglev:
Cheers! I appreciate your advice.

dhimmel:
`environment.yml` added in https://github.com/greenelab/library-access/pull/7

Issue #4, opened 2017-10-03
API calls in Python


dhimmel:
@publicus and I were discussion implementing the API calls in python.

Here are some helpful resources. First, most people use the [`requests`](http://docs.python-requests.org/en/master/) package for API calls, which is not builtin, but has a nice interface. Since this repository has many similarities to our Crossref API usage, you may want to checkout [`greenelab/crossref`](https://github.com/greenelab/crossref)... particularly the python and conda setup.

There are also lot's of additional packages that you may find useful such as [`backoff`](https://pypi.python.org/pypi/backoff) or [`requests_cache`](https://pypi.python.org/pypi/requests-cache). Basically, if you find yourself doing something highly involved, there may already be a python package for that!

Finally, if we ever want to do concurrent requests, which is probably not an immediate priority, I'd recommend `ThreadPoolExecutor` from [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html).

dhimmel:
If you want an on-disk solution to make sure you don't repeat queries, you could use `requests_cache`, [`sqlite3`](https://docs.python.org/3/library/sqlite3.html) directly, a [`shelve`](https://docs.python.org/3/library/shelve.html), or some XML database (like a MongoDB for XML, not sure if this exists).

In general, I think you will find it more manageable to have one big file with all the cached responses than separate files for each response.

jglev:
Thanks! I've not used `backoff`, `requests_cache`, or `shelves` before -- I appreciate the tips (and using this project to brush back up on Python, since R's been my daily language for a while). I'll look into those.

And if you prefer storing the XML in a database (I haven't read up on `shelve` yet, but I'd be comfortable using `sqlite3` immediately), I'm fine with implementing that.

dhimmel:
I think it'll be important to have a persistent cache, so you can avoid repeating any API calls. A `sqlite3` table would definitely get the job done. I could imaging the following columns:

+ `doi` (indexed)
+ `api_url` (or not if this should be private)
+ `timestamp` 
+ `xml_response`

Then when we've completed this analysis, we can upload the database file or export all of the XML snippets to a compressed text document for long-term archiving. The only downside I see to an SQLite database is that it will probably grow quite large in size, but this shouldn't be a problem.

jglev:
Yep, that schema sounds good to me!

[...]

Issue #1, opened 2017-09-25
Initial set of query DOIs from State of OA study


dhimmel:
A good starting set of DOIs would be those from the [State of OA study](https://doi.org/10.7287/peerj.preprints.3119v1), since there will be ~ 300,000 DOIs and will allow us to compare Library access to oaDOI and Sci-Hub access. I'll comment once I've generated this list.

dhimmel:
@publicus see [`state-of-oa-dois.tsv.xz`](https://github.com/greenelab/scihub/blob/4172526ac7433357b31790578ad6f59948b6db26/data/state-of-oa-dois.tsv.xz) with 290,120 DOIs that were assessed in the State of OA study. The `oadoi_color` columns indicates access status according to the oaDOI utility. For prototyping, you can work with a subset of these DOIs... but once we have info on all of them, that will be enough for several analyses.

jglev:
Great, thank you! 

I have two methods at this point that should work for resolving the DOIs to the University's holdings. The first way works right now, but would be harder to use for other universities/libraries, and only allows one DOI to be resolved at a time. The second uses an [OCLC](https://en.wikipedia.org/wiki/OCLC#Online_database:_WorldCat "Wikipedia: 'OCLC WorldCat Database'") API, and so could work for many more libraries, and should also allow multiple DOI lookup at once. However, that second method isn't working as its documentation describes it should. Thus, I'm waiting to hear back from the OCLC Support team before deciding which method to use.

jglev:
Also, @dhimmel, to confirm, from the `state-of-oa` dataset, you're suggesting just looking at the Closed (and possibly Bronze) DOIs, correct? (Of the 290,120 rows in the dataset, I see 182,804 that are Closed, and 226,690 that are either Closed or Bronze).

dhimmel:
> I'm waiting to hear back from the OCLC Support team before deciding which method to use.

Cool. You can always open Work In Progress pull requests before an analysis is complete if you'd like to start getting feedback.

> you're suggesting just looking at the Closed (and possibly Bronze) DOIs, correct?

I was thinking we'd query all 290,120 and filter at analysis time if desired. We don't yet know how the two methods will perform on these various types of articles. My guess is that Universities subscribe to subscription journals which will includes all articles that are closed, green, bronze, or hybrid. I agree that there is less point in querying gold articles, since libraries don't subscribe to these journals... but for the sake of comparison to the Sci-Hub analyses, let's query them anyways if possible.

[...]

PR #7, opened 2017-10-16
Adding api query script and example database from 10 DOIs

jglev:
This is a work-in-progress PR. The code is all functioning, but, on @dhimmel's suggestion, I'd appreciate ongoing feedback, given the Lab's [expectations](http://greenelab-onboarding.readthedocs.io/en/latest/index.html#document-code_review_checklist) for collaborators.

As soon as I get an in-progress thumbs-up on this, I'll start the UPenn downloader running on the dataset's DOIs.

## What this PR currently includes

1. Per @dhimmel [in a comment](https://github.com/greenelab/library-access/issues/6#issuecomment-336554051), I've included and manually edited the environment YAML file exported by Conda. I have not specified Channels specifically, because I was paring down the output of `conda env export`, which doesn't specify them except at the top.
    1. I've included `r-base` in this list. It isn't used yet, but is used in the Bayesian script I've written (and will add in a future commit or PR, once the downloader is underway).
1. Working code, written in Python3, for doing the following:
    1. Querying an OpenURL resolver for a given DOI.
        1. The URL and static parameters are easily configurable for each institution's unique OpenURL resolver. 
        1. Similarly, the function for querying the API is split into its own file and function, allowing it to be more easily modified for a given institution's specific needs.
    1. Parsing the response XML for a Yes/No indication of whether full-text access is available for that DOI.
        1. As above, the function for parsing the API response is split into its own file and function, allowing it to be more easily modified for a given institution's specific needs.
    1. Saving the response and the Yes/No indication into an SQLite database.
1. I expanded on the [database model](https://github.com/greenelab/library-access/issues/4#issuecomment-334181443) suggested by @dhimmel by creating a separate ID table for DOIs. This will allow us (or others) more easily to make the following two changes, if we ever want to:
    1. Import the original `tsv` dataset into the SQLite database.
    1. Rerun the same DOI more than once over time -- this is currently implemented (though we likely won't measure each DOI beyond one timepoint) and controlled by the variable `rerun_dois_that_are_already_in_database` in the configuration file.
1. Everything is, I think, [PEP8](https://www.python.org/dev/peps/pep-0008/) compliant (it is according to the linter built into Spyder).
1. The `state-of-oa-dois` dataset from @dhimmel [in this comment](https://github.com/greenelab/library-access/issues/1#issuecomment-331969935).
1. An example SQLite database resulting from running 10 of the `closed`-type DOIs from the dataset.
    1. Here are three queries to quickly explore this SQLite database, using [sqliteman](https://sourceforge.net/projects/sqliteman/) or whatever else:

        ```
        SELECT * FROM dois_table;  -- View the entire DOI table
        
        SELECT * FROM library_holdings_data;  -- View the entire DOI XML and full-text indicator table
        
        /* View the above tables, joined together */
        SELECT  
            dois_table.doi,  
            library_holdings_data.timestamp,
            library_holdings_data.xml_response,
            library_holdings_data.full_text_indicator
        FROM library_holdings_data
        JOIN dois_table
        ON library_holdings_data.doi_foreign_key = dois_table.database_id
        ```

    1. One important note about this dataset: It only indicates that full-text is available if it is *digitally* available. Put differently, it does *not* reflect whether a patron could physically go to a library shelf and get a hard copy of an article. This matches @dhimmel's original specification when we talked about it, but is nonetheless something I want to be clear about.

## Things that still need to be done, following the Lab's [checklist](http://greenelab-onboarding.readthedocs.io/en/latest/index.html#document-code_review_checklist)

1. Add metadata
    1. I haven't yet signed the files, because I'm unsure whether there's a standard way to do that in the Lab.
    1. I haven't yet added a commented docstring to each of the files.
    1. For the smaller functions, I haven't yet added a docstring.
1. Probably more, based on feedback here.


environment.yml
```diff
@@ -0,0 +1,21 @@
+name: UPenn_Holdings_Project
+channels:
+- anaconda-fusion
+- defaults
+dependencies:
+- beautifulsoup4=4.6.0
```
dhimmel: Let's specify the channel for each dependency, e.g. `anaconda::beautifulsoup4=4.6.0`. This makes it a lot easier to figure out where a package is coming from and makes things more deterministic.

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+    if static_api_request_parameters_dictionary is None:
+        static_api_request_parameters_dictionary = {}
+    if custom_user_agent_string is None:
+        custom_user_agent_string = ""
+
+    custom_api_query_header = {
+            'user-agent': custom_user_agent_string}
+
+    # Update the static api parameters to include the (dynamic) DOI:
+    api_request_parameters = static_api_request_parameters_dictionary
```
dhimmel: See that you may of run into one of python's worst gotchas. 

You may want to using `api_request_parameters={}` as the default parameter and then do:

```
api_request_parameters = api_request_parameters.copy()
```

This will make sure the function does not modify dictionaries passed by the user.

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+        item_doi,
+        api_base_url,
+        static_api_request_parameters_dictionary=None,
+        custom_user_agent_string=None):
+    """Given an item DOI (lacking the 'doi:/' prefix), query an OpenURL
+    resolver) and return XML from it."""
+
+    if static_api_request_parameters_dictionary is None:
+        static_api_request_parameters_dictionary = {}
+    if custom_user_agent_string is None:
```
dhimmel: What about just not setting this if not passed?

Like:

```python
headers = {}
if custom_user_agent_string:
   headers['user-agent'] = custom_user_agent_string
```

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+        custom_user_agent_string = ""
+
+    custom_api_query_header = {
+            'user-agent': custom_user_agent_string}
+
+    # Update the static api parameters to include the (dynamic) DOI:
+    api_request_parameters = static_api_request_parameters_dictionary
+    api_request_parameters.update({
+            'rft_id': 'info:doi/%s' % item_doi})
```
dhimmel: The customary syntax for adding a single item to a dictionary would be:

```python
api_request_parameters['rft_id'] = f'info:doi/{item_doi}'
```

See also the switch to Literal String Interpolation which is preferred over the `%` method.

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+
+    api_response = requests.get(
+            api_base_url,
+            params=api_request_parameters,
+            headers=custom_api_query_header)
+
+    class ErrorWithAPI(Exception):
```
dhimmel: This should be defined outside of the function at the module level

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+        pass
+
+    if api_response.status_code != 200:
+        raise ErrorWithAPI('Problem contacting API: {}'.format(
+                api_response.status_code))
```
dhimmel: So what happens in case of an error? Is that info logged anywhere?

library_management_system_downloader/evaluate_api_response_for_fulltext_indication.py
```diff
@@ -0,0 +1,40 @@
...
+    full_text_indicator_value = parseable_response_xml.find(
+            id='full_text_indicator')
+
+    logging.info(
+          'The value for the "full_text_indicator" key for doi "%(doi)s" is '
+          '"%(value)s".' % {
+                  'doi': doi_value,
+                  'value': full_text_indicator_value})
+
+    if(full_text_indicator_value is not None and
```
dhimmel: I'd probably do this:

```python
if full_text_indicator_value is None:
    return 0
return int(full_text_indicator_value.string == 'true')
```

library_management_system_downloader/downloader_configuration_file_TEMPLATE.py
```diff
@@ -0,0 +1,18 @@
+# These are parameters that stay the same from query to query (unlike the
```
dhimmel, line 1: This is a nice template. The non-template version should be ignored (added to `.gititnore`) right? To prevent accidental tracking.

Also you should add to the `.gitignore` any other files that are generated by are not intended to be tracked with git.

library_management_system_downloader/downloader_configuration_file_TEMPLATE.py
```diff
@@ -0,0 +1,18 @@
...
+        'project by Jane Doe of the University of ________. '
+        'I have put a threshold on the download in order '
+        'not to cause a problem with your servers. If this '
+        'download IS causing a problem, please feel free to email me at '
+        '________, or to call me directly at ________.')
+
+api_base_url = (
+            'https://example.com/openurl')
+
+tsv_dataset_location = '/path/to/dois.tsv'
```
dhimmel: Is this the query set or the output set? Let's use a variable name that makes that clear? Also would relative path be more appropriate (all paths within the repo should be relative)

library_management_system_downloader/evaluate_api_response_for_fulltext_indication.py
```diff
@@ -0,0 +1,40 @@
...
+       full_text_indicator_value.string == 'true'):
+        return 1
+    else:
+        return 0
+
+# =============================================================================
```
dhimmel: This comment should be moved to the docstring, deleted, or encoded as a unit test.

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,202 @@
+import logging
+import pandas as pd
+from sqlalchemy import create_engine
+from sqlalchemy import event
+from sqlalchemy import Table, MetaData, Column, String, DateTime, Integer, func
+from sqlalchemy import ForeignKey, distinct
+from sqlalchemy.orm import sessionmaker
+
+# =============================================================================
```
dhimmel, line 9: For these several line documentation inserts put them in triple quotes rather than line comments.

jglev:
@dhimmel, to confirm, is it ok to include the `state-of-oa-dois` dataset here? Do you have rights to release it CC0?

dhimmel (commented):
More comments on the way. Only part way done reviewing, but nice work so far. (especially for just picking python back up).

let's switch from the `%` method of string formatting to either `'string{}'.format()` or f-strings (https://cito.github.io/blog/f-strings/). f-strings are new in Python 3.6 and generally the best method.

environment.yml
jglev: I've done this in b66a9931c. Does that look correct to you? The only mention I've found so far re: which channel a package is from is by doing, e.g., `conda search pandas`, which lists `defaults` as the channel. So that's what I've used here. Does that jibe with your understanding?

jglev:
Thank you for the feedback (and in advance for whatever additional comments are forthcoming)!

As of 8d32f2f3dd, I've switched to using f-strings, which I hadn't heard about (thank you for that, as well). I'll be making further commits and comments above to track my progress tomorrow.

library_management_system_downloader/create_api_request.py
jglev: Sure! Implemented in b94df062cf.

library_management_system_downloader/create_api_request.py
jglev: Done in ea814de7e, as I understand.

library_management_system_downloader/create_api_request.py
jglev: Woah, that's confusing behavior of Python -- I hadn't realized that was a gotcha, but I get it now. Thank you for bringing it up.

I've made a change in ea814de7e for this -- does that match your suggestion to your satisfaction?

library_management_system_downloader/downloader_configuration_file_TEMPLATE.py
jglev: I've added `.gitignore` in 2e939. I'd had `.gitignore` listed in the file, and hadn't been tracking it until now.

dhimmel:
> @dhimmel, to confirm, is it ok to include the state-of-oa-dois dataset here? Do you have rights to release it CC0?

Certainly, the source [Zenodo release](https://doi.org/10.5281/zenodo.837902) is CC0. Now I don't think we should necessarily track `state-of-oa-dois.tsv.xz`. Instead, we can read it from its versioned URL:

```
https://github.com/greenelab/scihub/raw/4172526ac7433357b31790578ad6f59948b6db26/data/state-of-oa-dois.tsv.xz
```

In the above URL, you can switch between `raw` and `blob` (for webview)

.gitignore
```diff
@@ -0,0 +1,9 @@
+.directory
+__pycache__
+pull_request_Information.md
```
dhimmel: These seem like files you've created but don't necessarily want to be part of the repo. `.gitignore` should only reflect files that are created or contained in other instances of the repo. In other words, `pull_request_Information.md` is a file that only you will have. Therefore, it doesn't belong in the repo `.gitignore` because it means nothing to me.

On the other hand, `__pycache__` can get created by any user and is therefore appropriate to ignore. Another example of a file that applies to everyone would be a secrets file with a predetermined name.

environment.yml
```diff
@@ -0,0 +1,21 @@
+name: UPenn_Holdings_Project
+channels:
+- anaconda-fusion
+- defaults
+dependencies:
+- defaults::beautifulsoup4=4.6.0
```
dhimmel: `defaults` refers to the users local channel defaults. In that sense, it's not an actual channel, IIUC. Change these all to `anaconda` or `conda-forge` (you can mix or match).

environment.yml
```diff
@@ -0,0 +1,21 @@
+name: UPenn_Holdings_Project
+channels:
```
dhimmel: Remove channels (we will specify below)

environment.yml
```diff
@@ -0,0 +1,21 @@
...
+- defaults::pep8=1.7.0
+- defaults::pip=9.0.1
+- defaults::python=3.6.1
+- defaults::r-base=3.4.1
+- defaults::requests=2.14.2
+- defaults::spyder=3.1.4
+- defaults::sqlalchemy=1.1.9
+- defaults::sqlite=3.13.0
+- defaults::yaml=0.1.6
+- pip:
```
dhimmel: This `pip` section is good.

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if doi_foreign_key is None:
+        logging.info(f'Inserting DOI "{doi}" into the database...')
+        inserted_doi = dois_table_insert.execute(doi=doi)
+        # Get the foreign key we just inserted into the database:
+        doi_foreign_key = inserted_doi.inserted_primary_key[0]
+    else:
+        doi_foreign_key = doi_foreign_key[0]
+        logging.info(f'DOI "{doi}" is already in the database (foreign key '
+                     '"{doi_foreign_key}"), so not inserting it...')
```
dhimmel, line 128: ~~I think a REPLACE setup makes more sense. If someone has redone a query for an API that should be updated. Perhaps they want newer information. Users should avoid redoing queries by checking first whether a DOI has been queried, and if so not requerying it.~~

I crossed out above because I don't think I understand this part. Why is DOI a foreign key? Shouldn't it just be the primary key of the table? (this is a duplicate of my comment above)

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,201 @@
...
+            doi_foreign_key=doi_foreign_key,
+            xml_response=api_response_text,
+            full_text_indicator=full_text_indicator_for_doi)
+
+# =============================================================================
+# Import the list of DOIs
+# =============================================================================
+
+# Import the datset into
+doi_dataset = pd.read_csv(config.tsv_dataset_location, sep='\t', header=0)
```
dhimmel: `pd.read_table` defaults to `sep='\t'`. `header` not necessary.

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,201 @@
...
+# determine below whether to re-run DOIs that already have an answer in the
+# database.
+
+# =============================================================================
+# Query the API
+# =============================================================================
+
+for doi in list_of_dois[0:10]:
+    if (config.rerun_dois_that_are_already_in_database is not True and
+            is_doi_already_answered_in_database(doi) is True):
```
dhimmel: `is True` is superfluous.

.gitignore
```diff
@@ -0,0 +1,9 @@
...
+Datasets/State_of_OA/Not_Currently_Used
+Datasets/State_of_OA/state-of-oa-dois_CLEAN.tsv
+library_coverage_xml_and_fulltext_indicators.db*
```
dhimmel: This ignore should stay... eventually we'll switch it to only ignore the uncompressed db and a compressed db will be for archiving.

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,201 @@
...
+    cursor.execute("PRAGMA journal_mode=WAL")
+    cursor.close()
+
+# Create an sqlalchemy metadata schema:
+metadata = MetaData(sql_engine)
+
+# Create a schema for the table that will hold the API data:
+# The full_text_indicator column will be an integer in that 1 = fulltext
+# available and 0 = fulltext not available:
+dois_table = Table(
```
dhimmel, line 62: I think we should get rid of this table and just index the doi column of `library_holdings_table`.

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,201 @@
...
+        'dois_table', metadata,
+        Column('database_id', Integer, primary_key=True),
+        Column('doi', String, nullable=False, unique=True, index=True))
+
+library_holdings_table = Table(
+        'library_holdings_data', metadata,
+        Column('doi_foreign_key',
+               String, ForeignKey("dois_table.database_id"), nullable=False),
+        Column('timestamp', DateTime, nullable=False,
+               default=func.now()),  # Default is creation time
```
dhimmel, line 72: this is really cool!

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,201 @@
...
+        Column('xml_response', String, nullable=False),
+        Column('full_text_indicator', Integer))
```
dhimmel, line 74: I'm not sure `full_text_indicator` should be part of the database as opposed to created during a post-processing step. My thought was that the database would be solely for storing raw responses and avoiding duplicate API queries. The results could be processed later.

I don't have a super strong feeling here. If you go this way, you could save all XML processing for a future PR and we could just focus on API queries in this one.

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
+import backoff
+from ratelimit import rate_limited
+import requests
+import logging
```
dhimmel: The [PEP8 import convention](https://www.python.org/dev/peps/pep-0008/#imports) is to separate builtin, installed, and local imports via newlines:

```python
import logging

import backoff
from ratelimit import rate_limited
import requests
```

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+            'rft_id': f'info:doi/{item_doi}'})
+
+    api_response = requests.get(
+            api_base_url,
+            params=api_request_parameters,
+            headers=custom_api_query_header)
+
+    if api_response.status_code != 200:
+        raise ErrorWithAPI('Problem contacting API: {}'.format(
+                api_response.status_code))
```
dhimmel: Perhaps log the actual `api_response.text()` under `info` or `warning`? This may contain why the query failed.

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+
+# From the backoff documentation (https://pypi.python.org/pypi/backoff),
+# set maximum number of tries on a failed download:
+@backoff.on_exception(
+        backoff.expo,
+        requests.exceptions.RequestException,
+        max_tries=8)
+# From the ratelimit documentation (https://pypi.python.org/pypi/ratelimit),
+# set the rate limit (in seconds) for API calls:
+@rate_limited(0.2)
```
dhimmel: To make this more clear, can the comment say "max one query per five seconds"?

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+
+class ErrorWithAPI(Exception):
+    pass
+
+
+# From the backoff documentation (https://pypi.python.org/pypi/backoff),
+# set maximum number of tries on a failed download:
+@backoff.on_exception(
+        backoff.expo,
+        requests.exceptions.RequestException,
```
dhimmel, line 16: ~~Hm I'm not sure why backoff wouldn't be wanted for other exceptions like a `ConnectionError` or `Timeout`? I could even see it sometimes being helpful for any non 200-code response.~~

Now [I see](http://docs.python-requests.org/en/master/_modules/requests/exceptions/) that `RequestException` is a superclass.

library_management_system_downloader/create_api_request.py
dhimmel: We don't want non-200 responses to crash the program. They should perhaps be logged as warnings? In other words we need to know about these failures, but the querying should continue. Do you plan on handling these errors anywhere?

dhimmel (commented):
Read `Datasets/State_of_OA/state-of-oa-dois.tsv` from URL (pandas can do this). Not only does this prevent the file from being stored twice, it provides better provenance tracking.

Also for big files, we will want to use compression + Git LFS. We can save that for another PR. For now, don't track any datasets.

Generally, it's most efficient to keep PRs minimal in scope. I think we have narrowed down the scope of this PR to a unit that makes sense, although it's still a bit large. (more of a note for the future).

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+        max_tries=8)
+# From the ratelimit documentation (https://pypi.python.org/pypi/ratelimit),
+# set the rate limit (in seconds) for API calls:
+@rate_limited(0.2)
+def create_api_request(
+        item_doi,
+        api_base_url,
+        static_api_request_parameters_dictionary=None,
```
dhimmel: Part 1: do `api_request_parameters={}` here

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+        custom_user_agent_string=None):
+    """Given an item DOI (lacking the 'doi:/' prefix), query an OpenURL
+    resolver) and return XML from it."""
+
+    if static_api_request_parameters_dictionary is None:
```
dhimmel: Part 2: delete this line and the next

library_management_system_downloader/create_api_request.py
```diff
@@ -0,0 +1,51 @@
...
+        static_api_request_parameters_dictionary = {}
+
+    custom_api_query_header = {}
+    if custom_user_agent_string:
+        custom_api_query_header['user-agent'] = custom_user_agent_string
+
+    # Update the static api parameters to include the (dynamic) DOI:
+    api_request_parameters = static_api_request_parameters_dictionary.copy()
```
dhimmel: Part 3: replace `static_api_request_parameters_dictionary` with `api_request_parameters`.

Better to avoid using additional variable names in this case, since we will never want access to the original (pre-copy) dictionary.

.gitignore
jglev: That's actually why I wasn't tracking `.gitignore` before (because mine currently lists occasional scratchpad files I've been using for development).

Do you have a recommended approach for this workflow?

.gitignore
jglev: I'm leaving a note here that Write-Ahead Logging (WAL) for SQLite is enabled in the code, so when the code is run, two additional files are generated, `library_coverage_xml_and_fulltext_indicators.db-shm` and `library_coverage_xml_and_fulltext_indicators.db-wal` (hence the `*` in the `.gitignore` line above).

environment.yml
jglev: I've implemented this in 08802c7.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: This is my first project using `sqlalchemy`, and it's slick!

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: I see the argument both ways. My thought here is that the XML processing is minimal -- the only thing being done with the XML is to make a binary decision about whether full-text is available. Thus, since the `full_text_indicator` column just reflects what's in the XML, and not a big transformation of the XML, it makes sense to me to just store it alongside. Having said that, I also don't have super strong feelings about this, and the processing step is already its own function.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: The naming here might be unclear -- the DOI itself isn't a foreign key -- rather, `doi_foreign_key` refers to the Primary Key of the `dois` table. The `dois` table currently contains that Primary Key column, which is type `int`, and the DOI as a string. The lookup in this referenced chunk of code is just referencing the `dois` table -- if the DOI is already in that table, it won't get inserted again -- but either way, the download will continue, and will have its output placed into the data table, which is called `library_holdings_data`.

I originally wrote this with a single-table approach, but then came to think that this two-table approach offers more functionality, easier future expansion, and less need for error-catching, for these reasons:

- The main thing that this two-table approach allows is easily re-querying DOIs more than once -- perhaps to see changes over time, or to compare one library with another. With the current code, a configuration setting, `rerun_dois_that_are_already_in_database`, controls whether a DOI that's already been downloaded should be downloaded again. The two-table approach allows this, without requiring re-writing the DOI more than once in a table.
- An earlier draft I wrote did use `DOI` as the Primary Key. But since Primary Keys are unique, that approach wouldn't allow repeat downloads for the same DOI. And having that functionality is both potentially useful for others, and makes it easier for developing (because I can run the same download command several times to quickly check whether the database is working as expected).

Does that make more sense, as I'm explaining it here?

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: I'm fine taking out both. Having said that, since TSV is (in my experience) less-used than CSV, it makes sense to me to keep this explicit, so that anyone using a CSV might more quickly get the idea to change that (whether there or in a comment, I imagine that *I'd* like the reminder in the future). With that in mind, would you be ok with keeping it in?

library_management_system_downloader/downloader_configuration_file_TEMPLATE.py
jglev: I've made the variable name clearer in 06b0fc, by renaming it `input_tsv_dataset_location` and adding a comment above it.

jglev: And in 93478, have changed this to the relative path `'Datasets/State_of_OA/state-of-oa-dois.tsv'`

library_management_system_downloader/evaluate_api_response_for_fulltext_indication.py
jglev: I've encoded these as unittests in 468e1e.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: Done in e05bb2c.

environment.yml
jglev: Implemented in 08802c.

library_management_system_downloader/create_api_request.py
jglev: Done in 5bade9.

library_management_system_downloader/create_api_request.py
jglev: Sure, done in f8cf46c.
To confirm, though, did you mean "max five queries per one second"? (That's what I added in the above-referenced commit, and it seems to be correct from testing the code).

library_management_system_downloader/create_api_request.py
jglev: Do you mean to replace `raise ErrorWithAPI` with, e.g., `logging.warning()`? I'm fine with that.

I didn't plan on handling them elsewhere for now, as I actually thought it would be desirable for it to crash the program (since it would have up to 8 tries via `backoff`, a non-200 response would presumably mean more than a momentary connectivity interruption). The API isn't well-documented enough that I know what status codes might be returned, so I figured that if there's a problem, I should take a look at it manually, at least to start.

I've added the full response text (and switched to f-strings in this code chunk) in 658f7b4. If you would still like to see it not crash, though, let me know here, and I'd be fine to switch it to `logging.warning()`.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: Just noting that I responded re: this table below, [here](https://github.com/greenelab/library-access/pull/7#discussion_r145502339). We can certainly go back to a single-table setup, but I think that the two-table setup does have advantages (including not repeating DOI strings).

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: Removed in 71b93e.

library_management_system_downloader/create_api_request.py
jglev: So that line would read `api_request_parameters = api_request_parameters.copy()`? Or `api_request_parameters = api_request_parameters`?

Maybe I've been away from Python for too long, but I'm still not understanding this approach, sorry. (I'm not arguing against it; I'm just trying to understand it). My understanding is that setting a default parameter as `{}` is often undesirable since it can behave in ways that users don't expect. So I'm trying to understand why it *is* desirable here (since you suggested it).

Just making sure: The configuration variable is called `static_parameters_for_api` (not `static_api_request_parameters_dictionary`, which is what's in the code line referenced above). Do you mean for this line to read `api_request_parameters = static_parameters_for_api.copy()`?

Sorry if this seems (or is) dense; I appreciate your feedback on this approach.

.gitignore
dhimmel: Yes you should track a .gitignore with just the lines added that are relevant for everyone. Note that you can stage only certain lines of a file in git using `git add --patch`.

It's also fine for you to have local files that aren't ignored. I rarely do `git add --all`. Instead I just add the specific files that should be added (or all updated ones `git add --update`).

library_management_system_downloader/create_api_request.py
dhimmel: From looking at the [source code](https://github.com/tomasbasham/ratelimit/blob/1fe9ff9c18a97ff3e39731fd5d88f4e42a419310/ratelimit/__init__.py), I think the current implementation is saying run at most 0.2 times per second, which gets "clamped" to at most once per second. Documentation is poor, but I think you need to do:

```python
@rate_limited(period=1, every=5)
```

To say that the call can be made at most once within a five second period.

library_management_system_downloader/run_api_download_and_parse_results.py
dhimmel: > including not repeating DOI strings

What's the problem with repeating DOI strings? The extra disk space burden is irrelevant compared to the size the of the XML responses. IMO creating an internal set of identifiers is much more likely to cause a problem.

library_management_system_downloader/run_api_download_and_parse_results.py
dhimmel: > to compare one library with another

If we want to do this, we should add a library column. You could then use (`library`, `doi`, `timestamp`) as the primary key. Note primary keys are not necessary in this single table approach... they just would help us enforce uniqueness constrains, which may or may not be desirable?

> Primary Keys are unique, that approach wouldn't allow repeat downloads for the same DOI

Are there any benefits that wouldn't be solved by placing an index on the DOI column?

library_management_system_downloader/run_api_download_and_parse_results.py
dhimmel: The point is you're reading a TSV so you should use the function that defaults to tab-separation.

For the crossref and scihub repos, we're using TSV as the tabular plain text format. I'd recommend continuing with TSV output (always with column names) for this repo. Many of our fields may contain commas, but will rarely contain tabs.

library_management_system_downloader/create_api_request.py
dhimmel: This line would become

```python
api_request_parameters = api_request_parameters.copy()
```

My suggestion is to use one variable name to refer to the dictionary prior to copying and the dictionary post copying. They will be different objects in memory.

However, performing the copy ensures that you don't edit a user-provided dict or the default parameter empty dict.

According to my suggestion `static_api_request_parameters_dictionary` will no longer appear in the code.

library_management_system_downloader/create_api_request.py
dhimmel: If you want "max five queries per one second" I think you want `@rate_limited(5)` which is `@rate_limited(period=5, every=1)`

.gitignore
jglev: I've implemented reading the TSV file from the URL you provided in e135948.

jglev: I've pared down the `.gitignore` file per your specifications in 2b8e44.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: There's no problem with repeating DOI strings, nor is there, I think, with not repeating them. If repeat downloads are to be allowed (which isn't something we talked about originally, but does make sense to me and makes development easier, as I mentioned in a [comment above](https://github.com/greenelab/library-access/pull/7#discussion_r145502339) (linking to it because there are a lot of comments in this thread at this point), DOI can't be the Primary Key of the table. In that case, we could have one of the following:

1. A table with a throwaway Primary Key and an indexed DOI column: This isn't *wrong;* it just violates [convention](https://en.wikipedia.org/wiki/First_normal_form), so I didn't use it here.
2. A table without a Primary Key: As above, not *wrong* (whatever that would mean here), just [unconventional](https://stackoverflow.com/questions/840162/should-each-and-every-table-have-a-primary-key), and potentially making it harder for others to re-tool the work for any expansions they want to do. My understanding from your [comment below](https://github.com/greenelab/library-access/pull/7#discussion_r146052018) is that this is what you're suggesting; is that correct?

Having thought on this, I don't think that there's a "correct" answer -- the current implementation works, and, to my understanding, follows some schema best practices. Having said that, I don't disagree with what you've written; I see it as basically an aesthetic choice. The design works now and would also work the other way. With that in mind, my goal is to have this PR accepted, which I know is at your discretion. As the project lead, do you have strong feelings about this? In suggesting the one-table solution, are you *requiring* the one-table solution?

jglev: (I hope that that comment comes across as fully respectfully as I mean for it to; I think I'm just confused about where you're requiring changes vs. just suggesting them / opening conversation about design decisions in your comments.)

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: I don't envision *us* comparing Libraries, but I do see it making sense if others want not only to view the code but to re-use it. So I can go ahead and add a `library` column (or foreign key, as well as a `doi` column, depending on your response to [a comment I just made above](https://github.com/greenelab/library-access/pull/7#discussion_r147198487)).

And I agree that, if DOI isn't serving as Primary Key, placing an index on it would be desirable, and would make the table work fine.

library_management_system_downloader/create_api_request.py
jglev: Implemented in 6d38bb6.

library_management_system_downloader/create_api_request.py
jglev: I confirmed the two-argument behavior you suggested just now with a test, and have updated the comment and syntax (which now uses explicit arguments, and the second (`every`) argument, in f81cd152a.

The [pypi](https://pypi.python.org/pypi/ratelimit) page doesn't mention the second argument, which is why I missed it!

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: I didn't notice in your initial comment that you were suggesting replacing `read_csv` with `read_table`. I've done that in 930fc14, removed the default arguments, and added a comment clarifying that a TSV is expected.

library_management_system_downloader/create_api_request.py
jglev: I appreciate your patience and taking the time to explain; thank you. Implemented in 6d6c746a6.

library_management_system_downloader/evaluate_api_response_for_fulltext_indication.py
jglev: I'm fine with this; implemented in 4180e5.

library_management_system_downloader/create_api_request.py
jglev: Just a note that I've answered this in a comment [below](https://github.com/greenelab/library-access/pull/7#discussion_r145510129).

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def is_doi_already_answered_in_database(
+        doi,
+        sqlalchemy_session=sql_session):
+    existing_doi_with_fulltext_anwer = sql_session.query(distinct(
+        dois_table.c.doi)).join(library_holdings_table).filter(
+            dois_table.c.doi == doi).filter(
+                    library_holdings_table.c.full_text_indicator.isnot(
+                            None)).first()
```
dhimmel: The newlines here make it a bit hard to read.

I would go with the one operation per line style that is becoming popular in the tidyverse and pandas for chaining:

```python
existing_doi_with_fulltext_anwer = (
    sql_session.query(distinct(dois_table.c.doi))
    .join(library_holdings_table)
    .filter(dois_table.c.doi == doi)
    .filter(library_holdings_table.c.full_text_indicator.isnot(None))
    .first()
)
```

If some of these lines exceed the max characters per line that's fine IMO. The alternative to have it be hard to read is worse. If you want to suppress the style warning you can always follow the line with `  # noqa: E501` to suppress  "line too long"

dhimmel: `anwer`? mean `answer`

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if existing_doi_with_fulltext_anwer is None:
```
dhimmel: Would `return bool(existing_doi_with_fulltext_anwer)` work here?

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,202 @@
...
+# Import the datset
+# Note that read_table defaults to using TSV (tab-separated) format and using
+# the first row as a header row.
+doi_dataset = pd.read_table(
+        config.input_tsv_dataset_location,
+        compression='infer')  # Allow Pandas to read compressed files.
+
+# set() will get unique values in the list:
+list_of_dois = list(set(
+        doi_dataset[doi_dataset['oadoi_color'] == 'closed']['doi']))
```
dhimmel: I may do

```python
sorted(set(doi_dataset.query("oadoi_color == 'closed'").doi))
```

Sorted will give you a list but ensure deterministic ordering

library_management_system_downloader/evaluate_api_response_for_fulltext_indication.py
```diff
@@ -0,0 +1,47 @@
...
+        return 0
+    return int(full_text_indicator_value.string == 'true')  # This will
+    # evaluate to either 1 or 0.
+
+# =============================================================================
+# Function examples / tests
+# =============================================================================
+
+
+class TestFulltextIndication(unittest.TestCase):
```
dhimmel: We should use pytest for testing... but we can save that for a later PR.

environment.yml
```diff
@@ -0,0 +1,18 @@
...
+- anaconda::python=3.6.1
+- anaconda::r-base=3.4.1
+- anaconda::requests=2.14.2
+- anaconda::spyder=3.1.4
+- anaconda::sqlalchemy=1.1.9
+- anaconda::sqlite=3.13.0
+- anaconda::yaml=0.1.6
+- pip:
+  - backoff==1.4.3
+  - ipython-genutils==0.2.0
```
dhimmel: What's this for? [PyPI package](https://pypi.python.org/pypi/ipython_genutils) says:

> Pretend this doesn’t exist. Nobody should use it.

library_management_system_downloader/run_api_download_and_parse_results.py
dhimmel: > where you're requiring changes vs. just suggesting them / opening conversation about design decisions in your comments

All the comments are suggestions. If you disagree then let me know. Of course, sometimes I'll make comments that are wrong. The current level of discussion is good.

Regarding the database design (one v two tables), I'll leave it up to you. I was mostly envisioning the database as an indexed on-disk way to store the API queries. But if you're envisioning doing more with the DB than a more robust design may make sense.

dhimmel (commented):
untrack `library_coverage_xml_and_fulltext_indicators.db`.

After this PR is merged, we can do a small PR to track the output (db file and a TSV of `DOI` and `full_text_indicator` for external apps).

environment.yml
jglev: Removed in b3e7ebca. Conda added it when I originally generated the `environment` file, and I didn't notice it to delete it.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: Fixed the spelling error and changed the indentation in aad46c7a.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: I think so -- implemented in 7f2741.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: That sounds useful to me, too -- implemented in 465880.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: ...and fixed that commit in 7d6ade0, where I also added a new config. setting, `oadoi_values_to_limit_search_to`, to streamline the process of choosing which DOIs to download (this was previously hardcoded in the script itself).

jglev:
As @dhimmel suggested [here](https://github.com/greenelab/library-access/pull/7#pullrequestreview-72890801), I stopped tracking the database in ab6541ba50.

In f1d2d67, I added a new config. variable to slice the list of DOIs -- previously, it was hard-coded to just take the first 10 (as a test set). So with a variable (which when set to `None`, will download everything), the PR can now be used without having to go back and change that slice in the code itself.

dhimmel:
@publicus cool. Let me know when its ready for my review... i.e. no more planned commits.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: Following your comment [above](https://github.com/greenelab/library-access/pull/7#discussion_r147771042),

> Regarding the database design (one v two tables), I'll leave it up to you. I was mostly envisioning the database as an indexed on-disk way to store the API queries. But if you're envisioning doing more with the DB than a more robust design may make sense.

I do take your point, and think that ["You Aren't Gonna Need It"](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) is generally a worthwhile philosophy to adhere to. In this particular case, I do prefer the two-table design. Not because I think the one-table design is worse, but because I can imagine myself playing with the data in a way that would make the two-table design useful. For example, if we did bump this up to querying millions of DOIs, as you originally suggested, it seems to me that it could be useful in that case to import the State of OA dataset (with the OA "color" column) into that second table, to allow faster subsetting of DOI results based on color. That's something that we don't need to do right now, but could easily want to do. And having the second table in place would make that quicker to do.

Had I had "YAGNI" in mind, I probably wouldn't have made the second table. But since I have, and it's a quick `JOIN` to get it into a single table, it makes more sense to me to leave it with two tables. Does that sound reasonable to you, too?

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: ...Annnd fixed a typo in 2b00485.
I found that Spyder/iPython was erroring out with of an indentation error until I added the slashes at the end of each line here, so that's why I added them.

jglev:
I've gone through the PR discussion above, and think that all of your comments to date have been answered at this point. Does that look right to you, too?

.gitignore
```diff
@@ -0,0 +1,4 @@
+.directory
+__pycache__
+library_coverage_xml_and_fulltext_indicators.db*
```
dhimmel, line 3: `library_coverage_xml_and_fulltext_indicators.db` still part of the diff. Can you run:

```
git rm --cached library_coverage_xml_and_fulltext_indicators.db
```

environment.yml
```diff
@@ -0,0 +1,17 @@
+name: UPenn_Holdings_Project
```
dhimmel: I think it makes sense to use the same name for the environment as the repo (`library-access`), just so the relationship is more obvious.

library_management_system_downloader/run_api_download_and_parse_results.py
```diff
@@ -0,0 +1,210 @@
...
+# Note that read_table defaults to using TSV (tab-separated) format and using
+# the first row as a header row.
+doi_dataset = pd.read_table(
+        config.input_tsv_dataset_location,
+        compression='infer')  # Allow Pandas to read compressed files.
+
+# set() will get unique values in the list:
+list_of_dois = sorted(set(
+        doi_dataset[
+                doi_dataset['oadoi_color']
```
dhimmel, line 151: It would be better for this script to consume a list of DOIs rather and not hardcode this very specific structure of a single input. This change will be necessary if we query DOIs beyond the oaDOI dataset. Let's defer making a fix for later.

dhimmel (commented):
Just two small things, then I plan on merging.

library_management_system_downloader/evaluate_api_response_for_fulltext_indication.py
jglev: I had a few extra minutes today, and this seemed small, so I added a tests file for the function, in 6f2653. Is that how you prefer it?

library_management_system_downloader/tests_evaluate_api_response_for_fulltext_indication.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_fulltext_indication():
+    assert to_test.fulltext_indication(
+            '<key id="full_text_indicator">true</key>') == 1
+
+    assert to_test.fulltext_indication('tester') == 0
+
+# Run the tests:
+test_fulltext_indication()
```
dhimmel: This line not needed. pytest will run the function automatically because it starts with `test_`!

Not sure if `tests_evaluate_api_response_for_fulltext_indication.py` has to be renamed to `test_evaluate_api_response_for_fulltext_indication.py` for test detection.

jglev: From the [documentation](https://pytest.readthedocs.io/en/reorganize-docs/new-docs/user/naming_conventions.html), it looks like it does need to be `test_`. I renamed it, and deleted the unneeded line, in 2983061.

.gitignore
jglev: Done!

environment.yml
jglev: Sure, no problem. Done in d035fe6.

library_management_system_downloader/run_api_download_and_parse_results.py
jglev: I take your point. That sounds good to me.

dhimmel (approved):

jglev:
Thank you for all of your comments and review! I'll get this running today!

dhimmel:
> Thank you for all of your comments and review! I'll get this running today!

Make sure you start your subsequent work on top of the current `greenelab:master`. You need to be building off of the commit we just created by the squash merge.

jglev:
Will do!

PR #9, opened 2017-11-02
Enable Travis CI for testing

dhimmel:
@publicus can you review this.

The purpose is to use TravisCI to automate the testing. This way if a PR breaks the environment or tests, we'll know it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .directory
 __pycache__
+.cache
+
 library_coverage_xml_and_fulltext_indicators.db*
 downloader_configuration_file.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+dist: trusty
+sudo: false
+language: generic
+before_install:
+  - wget https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh
+    --output-document miniconda.sh
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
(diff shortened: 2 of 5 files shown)


jglev (commented):
Everything here looks good to me.

I haven't set up Travis before, but what you've put together looks sensible to me after reading the documentation.

One thing to confirm: Do we need to add `.` to the `PATH` / `PYTHONPATH` to the Travis setup? (I need to use `PYTHONPATH=.` when running the download script from the command line, so I'm asking here.)

The edits to the Readme read well to me. And `@pytest.mark.parametrize` is cool! I hadn't seen that before.

jglev:
I'll click Approve on this as soon as you confirm the one question I brought up.

dhimmel:
CI is nice because it makes it easy to catch things like the environment issue from #10 (see the corresponding [build failure](https://travis-ci.org/greenelab/library-access/builds/296343355#L678)).

> Do we need to add . to the PATH / PYTHONPATH to the Travis setup?

At the moment this isn't required for the existing tests. I'm not sure modifying your `PATH` is the right solution. We should try to make it so you can run your commands from the repo and they work based on relative links.

Perhaps save this issue for another PR that actually runs the scripts?

jglev:
Sure!

jglev:
I don't see a button at this point to do an official "Approve" step for review. To clarify, should I be seeing that, or can you just merge at this point?

dhimmel:
> I don't see a button at this point to do an official "Approve" step for review. To clarify, should I be seeing that, or can you just merge at this point?

Under files changed, click review changes. Then you can select Approve. I can merge now, but might as well get the official approval as best practice.

jglev (approved):

jglev:
Thanks!

PR #10, opened 2017-11-02
Added dependency for download script. 

jglev:
I just tried using the download script on a fresh computer (to run the downloads on my desktop vs. my laptop), and it turns out `lxml` needs to be listed, as well. It was installed on my laptop, but I didn't remember / notice it when paring down the `environment.yml` file that Conda automatically initially created.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,3 +15,4 @@ dependencies:
 - pip:
   - backoff==1.4.3
   - ratelimit==1.4.1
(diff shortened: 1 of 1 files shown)


dhimmel:
@publicus please see #9 ... I already have a PR open to fix this.

jglev:
Ah, sorry; I hadn't seen that. I'm reviewing #9 now...

Issue #11, opened 2017-11-02
Merging the downloaded data


jglev:
Following the merge/close of PR #7, I have the downloader running now.

While it runs, @dhimmel, perhaps we could discuss what you'd like done with it when it's ready? (Where it should go, whether any transformations / compression needs to be run on it, etc.)

dhimmel:
So there are two primary data outputs:

1. `library_coverage_xml_and_fulltext_indicators.db`
2. a TSV which gets extracted from `library_coverage_xml_and_fulltext_indicators.db`

In general I like having a `data` directory to host data files, but wherever they'll fit best is fine. We should track an xz-compressed version of `library_coverage_xml_and_fulltext_indicators.db` using Git LFS. Unless we think the xz-compressed db will exceed 2 GB. Then we should not track it using git because GitHub will reject the file.

The next question is whether to compress the TSV. The benefit to compressing is smaller file sizes. With compression, I'd suggest enabling git LFS for that file. The benefit to not compressing is that we can get line-by-line diffs as the file changes. Being able to `git diff` the TSV can help catch errors and provide feedback on updates. We should compress the TSV if it will exceed 50 MB.

dhimmel:
> I have the downloader running now.

On how many DOIs. The intial PR will probably be easiest with few DOIs. Then we can ramp up

jglev:
> On how many DOIs.

Since a full download will take several days, my plan is to just leave it running while I'm at work, until I have the full dataset (I was starting with the `closed` DOIs, but also did some downloads just now to make sure that it works if I list all of the categories in the config. file). I can always just commit a few records in a PR initially if that makes it easier, though.

> Unless we think the xz-compressed db will exceed 2 GB.

I've downloaded 4,345 DOIs so far, and the database is 27.8 MB. Assuming (perhaps naively, but also, I expect, conservatively) that storage will scale up linearly with more records, that's `27.8/4345 = 0.0064` MB per record. With ~300,000 records, that's `0.006*300,000 = 1800` MB of storage space for slightly more than the whole dataset (which has closer to 290,000 records). Thus, I expect we won't go over the 2 GB limit you mentioned.

In the `data` directory you suggest, would you like just the output `tsv` in there, or the database, too?

dhimmel:
> 0.0064 MB per record

Was this with or without compression? I imagine compression will cut down the size considerably. In the longrun, I'd like a solution that can accommodate all ~80 million DOIs.

[...]

PR #12, opened 2017-11-03
Extract tsv and compress database copy

jglev:
This follows our discussion at #11. @dhimmel, I *think* I've followed what you meant, but I'm not sure (I feel dense, but I suspect I'm just not reading your comments correctly). The code here does the following:

1. Take the database and create a TSV from it, with DOI, Timestamp, and Full_Text_Indicator as columns. It seemed unnecessary to me to include the full XML response, but it would be minor to add it. Data are sorted by timestamp and then (albeit probably unnecessarily) by DOI, so that the data will stay in the same order as we add to them over time.
    The TSV is currently saved with `xz` compression (I originally mistakenly used `gzip`), which, as I understand, is what you requested (is that correct?).
    This is currently saved in the `data` directory.
1. Take the database and create a `lzma`-compressed copy of it, in the `data` directory.
    This file is tracked with Git LFS.
1. Add a primary key to the main database table, because it turns out it's difficult to get `sqlalchemy` to use a table that doesn't have a primary key (documentation reference is in the code, and [here](http://docs.sqlalchemy.org/en/latest/faq/ormconfiguration.html#faq-mapper-primary-key) (cf [the note at the bottom of this section](http://docs.sqlalchemy.org/en/latest/orm/extensions/automap.html#basic-use))).
1. Fix a small bug I found in the downloader code (the config setting for selecting a test set of rows wasn't being honored).


.gitattributes
```diff
@@ -0,0 +1 @@
+data/library_coverage_xml_and_fulltext_indicators.db.xz filter=lfs diff=lfs merge=lfs -text
```
dhimmel: I would probably just do `*.xz` so any xz compressed files get LFS tracked.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
```diff
@@ -0,0 +1,102 @@
...
+
+joined_dataset = pd.read_sql(
+            dataset_join_query.statement,
+            dataset_join_query.session.bind)
+
+# Sort the dataset by DOI and timestamp, so that it will always be in the
+# same order:
+joined_dataset.sort_values(
+        by=["timestamp", "doi"],
+        inplace=True)
```
dhimmel: I'm thinking that me way want a more minimalist output TSV, which would just have a `doi` and `fulltext` column. When there are multiple timepoints for a DOI, it would report latest.

If you would like a TSV that includes the timestamps, let's just make two TSVs.

dhimmel: I agree that the TSVs should not have a field for the whole XML response.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
```diff
@@ -0,0 +1,102 @@
...
+# Create a configured "session":
+Session = sessionmaker(bind=sql_engine)
+sql_session = Session()
+
+# =============================================================================
+# Create the data directory if it doesn't already exist
+# =============================================================================
+
+if not os.path.exists(data_directory):
+    os.makedirs(data_directory)
```
dhimmel: Not essential but you may enjoy `pathlib`, which is now builtin and is great for doing things with paths.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
```diff
@@ -0,0 +1,102 @@
...
+database_output_path = os.path.join(
+        data_directory, compressed_database_copy_name)
+
+logging.info(
+        f'Creating a copy of the database under "'
+        f'{database_output_path}"...')
+
+with open(sqlite_database_location, "rb") as database_file:
+    with lzma.open(database_output_path, "wb") as output_file:
+        output_file.writelines(database_file)
```
dhimmel: Are you sure this works? I'm a little nervous the `writelines` approach may break something.

It may be preferable to use the `xz` command line program here. You could even do it with a [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) call. `xz` [can be added](https://anaconda.org/conda-forge/xz) to the environment, if we want to ensure everyone has the utility. The two things I worry about with the above command are:

1. Is it actually xz compressing the file (could it split the bytes for a windows newline and write a unix newline)?
2. Will it be memory efficient or will large chunks of the file be loaded into memory at once?

dhimmel (commented):
Nice.

.gitattributes
jglev: Sure, I'm fine with that. To confirm, though, that would also include the TSV file, which currently is also saved with a `.tsv.xz` extension (saving it without was causing LibreOffice to try to open it directly on my system, resulting in garbled-looking data). I remember you mentioning that the data file itself shouldn't use LFS. Is it ok with you to use LFS for both?

dhimmel: If a file is compressed, it should have the extension for its compression. If this is the case, not sure why LibreOffice would try to open your XZ (`*.tsv.xz`) compressed file.

If a file is compressed, it should be tracked using git LFS.

If you want line-by-line diffing for a file, it should not be compressed but rather plain text.

Does this answer your questions?

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: Thanks! Implemented in fb053e5.

.gitattributes
jglev: I meant that I was originally compressing the TSV but not adding the `.xz` extension, and LibreOffice would try to open that. So I added the `.xz` extension, which now correctly gets opened in my system's archive manager program, which shows the uncompressed file that LibreOffice can successfully open.

So yes, that does answer my question: since the TSV *is* compressed, I'll track it with LFS, too, using the edit you suggested.

jglev: Implemented in a09dc9.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: I haven't tested it manually yet, but I've drafted what I think should work in c605a7366. Does this look correct to you, too?

jglev: (The draft returns just the `doi` and `fulltext` columns, but does so for only the latest timestamp for each DOI.)

jglev: Updated annotations for this in b8a0a3674.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: So, it definitely does work, at least on my system -- I tested it manually several times, where I'd download some records, run this script, open up the now-compressed database copy using `sqliteman` and compare its number of records to those in the output `tsv` and the original, uncompressed database, then download some more records and repeat the process.

But, I'm not sure of its efficiency, which it seems you're primarily asking about. I'm open to using the command-line `xz` command. The current approach of this whole chunk of code does follow the canonical `lzma` example from the [library's documentation](https://docs.python.org/3/library/lzma.html#examples), though, as I understand. `open()` [does have a `buffering` argument](https://docs.python.org/3.4/library/functions.html#open), which can be used

> to indicate the size in bytes of a fixed-size chunk buffer. When no buffering argument is given, the default buffering policy works as follows:
>    - Binary files are buffered in fixed-size chunks; the size of the buffer is chosen using a heuristic trying to determine the underlying device’s “block size” and falling back on io.DEFAULT_BUFFER_SIZE. On many systems, the buffer will typically be 4096 or 8192 bytes long.

Would `write()` be a preferable choice over `writelines()` here?

I'm willing to defer to your greater expertise with Python -- given the documentation I've cited above, which approach do you prefer?

.gitattributes
dhimmel: @publicus, when writing commit hashes in GitHub, I'd recommend using at least 7 characters. For example, `a09dc99 versus a09dc9` is written below:

a09dc99 versus a09dc9. GitHub needs at least 7 characters to auto-link.

> Implemented in a09dc9

Notices such as above are helpful, although not essential... so do it only if it helps you organize things.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
dhimmel: I'd love for a pure python implementation of memory-efficient `xz` compression a file. However, I don't see that as an [example in the docs](https://docs.python.org/3/library/lzma.html#examples). The examples don't include a concurrent read-compress-write example.

However, there is an [example](https://docs.python.org/3/library/gzip.html) in the `gzip` docs:

```python
# Example of how to GZIP compress an existing file:
import gzip
import shutil
with open('/home/joe/file.txt', 'rb') as f_in:
    with gzip.open('/home/joe/file.txt.gz', 'wb') as f_out:
        shutil.copyfileobj(f_in, f_out)
```

You should be able to apply this method but use `lzma` instead. I recommend making a `xz_compress` function, in case we want to call this multiple times.

I'm not convinced the current method you have is memory efficient and will never corrupt a file.

dhimmel: You can group the with statements:

```python
with open(path_in, 'rb') as f_in, lzma.open(path_out, 'wb') as f_out:
    shutil.copyfileobj(f_in, f_out)
```

dhimmel:
This is looking good:

```sh
curl --location --silent https://github.com/publicus/library-access/raw/b8a0a36746395ab51a6426f56908b50269f28982/data/library_coverage_xml_and_fulltext_indicators.tsv.xz | xzcat | head
```

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
```diff
@@ -0,0 +1,119 @@
...
+        dois_table.doi,
+        # library_holdings_data.timestamp,  # Uncomment to also get the
+        # timestamp of the DOI fulltext query
+        # library_holdings_data.xml_response,  # Uncomment to also get the
+        # full XML response.
+        library_holdings_data.full_text_indicator) \
+        .join(library_holdings_data) \
+        .group_by(dois_table.doi) \
+        .filter(library_holdings_data.primary_key.in_(
+                dataset_max_timestamp_for_each_doi_query))
```
dhimmel, line 84: I expect that there is a way to sort by date within a group and just take the first row. Not a huge concern for me but this method seems a bit overkill?

jglev:
I'm sorry for my radio silence -- I'm prepping to teach a big workshop, and am looking forward to implementing/answering your suggestions! If I dive back into this on Monday, @dhimmel, will that work with your timeline?

dhimmel:
> If I dive back into this on Monday will that work with your timeline?

That's great! I'm currently traveling in Europe and get back Monday.

jglev:
Perfect. Safe travels!

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: As of 4dbfd6, I've done the following:

1. Implemented the `gzip`-documentation example you provided, but using `lzma` compression.
2. Grouped the `with` statements.
3. Wrapped the database compression into a function (called `xz_compress_a_file`)
4. Manually confirmed that the above works with the example database, by running the function on it, then opening it with `sqliteman` and running `SELECT COUNT(*) FROM dois_table` to confirm that it has the correct number of rows (currently 54).

Thanks for the suggestions!

dhimmel: Did you push `4dbfd6`?

dhimmel: @publicus I think you forgot to push `4dbfd6`

jglev: Thanks for the poke. Right you are -- I did `git push origin master`, forgetting I was on a different branch. It's [here](https://github.com/greenelab/library-access/pull/12/commits/4dbfd60223a40e2e794fd516b46f61795833816b) now : )

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: I need to sit down tomorrow and experiment with this -- I didn't know that SQL will allow you to sort first and then do a `GROUP BY` on those sorted rows in one query (am I reading your suggestion correctly, that that's what you're suggesting?). This approach translates out of `sqlalchemy` into a standard approach I've seen, to use a subquery to get the IDs one wants first, and then to return just those. Having said that, line 81 is possibly superfluous(? I'm not sure yet).

I think that if we loaded the whole database into a `pandas` dataframe, it would be more straightforward to do what you suggested. But I remember you mentioning earlier that you'd like the code to work at larger scales, too. So keeping things within the SQLite database and then returning just the already-filtered rows seems intuitively to me to be more in line with that goal. Does that all jibe with your understanding?

dhimmel: My suggestion (which I think you understand) is:

1. sort by date
2. groupby
3. take first row within group

Of course, 1 and 2 can likely be swapped. My intuition is that this method is possible in SQL / sqlalchemy and more straightforward and efficient than the current approach. However, if my suggestion seems difficult and the current approach is working, then we can continue with it.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
```diff
@@ -0,0 +1,146 @@
...
+
+# =============================================================================
+# Save a compressed copy of the database
+# =============================================================================
+
+
+def xz_compress_a_file(
+        input_file_location,
+        output_file_location,
+        is_the_file_binary=True):
```
dhimmel: I don't understand the need for `is_the_file_binary`... you always treat a file as binary when compressing it?

.gitignore
```diff
@@ -1,6 +1,5 @@
 .directory
 __pycache__
 .cache
-
-library_coverage_xml_and_fulltext_indicators.db*
+./library_coverage_xml_and_fulltext_indicators.db*
```
dhimmel, line 4: This seems like it would ignore `data/library_coverage_xml_and_fulltext_indicators.db.xz` but its currently tracked?

dhimmel (commented):
Almost there.

Did we decide whether this PR will contain the output data or not? Perhaps we should not track the outputs in this PR so we can merge it quickly. Then in the next PR, you can query all State of OA DOIs and track the outputs?

Would be great if we could get the output TSVs in ~ a week, since final submission is [now in sight](https://github.com/greenelab/scihub-manuscript/issues/33).

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
dhimmel: Also this argument is entirely non-functional at the moment. [`lzma`](https://docs.python.org/3/library/lzma.html) defaults to binary mode... `r` is equivalent to `rb` and `w` is equivalent to `wb`.

dhimmel: Since memory-efficient on-disk compression took us a while to figure out, I [opened an issue](https://bugs.python.org/issue32160) on Python's tracker. So maybe one day we can get our solution added to the official Python!

.gitignore
jglev: `data/library_coverage_xml_and_fulltext_indicators.db.xz` is tracked because it's listed in the [`.gitattributes`](https://github.com/greenelab/library-access/pull/12/files#diff-fc723d30b02a4cca7a534518111c1a66) file, for use with Git LFS, as I remember/understand.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: There's a bug here -- fixed in 0a720d5. Does this make more sense? I meant to only use `is_the_file_binary` for the input file. I can't find where we originally discussed it, but as I remember, we talked about wrapping the `lzma` code into a generic compression function, so I added it, since it was quick and would allow use on non-binary files. I'm fine taking it out, if you'd prefer; it's a minor change.

jglev: (In the commit before that, 8d554f9, I also replaced some wrong variable names, that were vestigial from before I'd wrapped the code into a function.)

jglev:
Also, I've noted the [timeline](https://github.com/greenelab/library-access/pull/12#pullrequestreview-79638756) you suggested!

jglev:
In response to your [question about the scope of this PR](https://github.com/greenelab/library-access/pull/12#pullrequestreview-79638756), I don't remember definitively (does GitHub hide comments on old code? I'm having trouble finding correspondence that I vaguely remember us having), but I think we agreed to track output in this PR. With that said, I feel fine switching tack and just tracking code here! I defer to your decision; if we do track just code, do I need to do anything on my end beyond deleting the example files and making a final commit on this branch?

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: Thanks for clarifying, so that I'm sure we're on the same page! : )

I remember more clearly now why I used the approach I did. I'm writing all this out because I initially agreed with your intuition that this can be simpler, but then started writing out the steps I'd need to take, and ended up thinking that the current approach is actually *necessary* (though I'd be interested to be shown to be wrong!). I want to make sure my thinking is clear about it (so that we can both review it, and so that it's easier to remember if needed in the future):

As I understand, SQL requires that `GROUP BY` statements come before `ORDER BY` statements. Since we want to be sure of the order of the rows before grouping them, we'd need to use a subquery to order them first, like this:
```sql
SELECT ...
FROM (
    SELECT ...
    FROM ...
    ORDER BY ...
)
GROUP BY ...
```
which is basically what the `sqlalchemy` statement does: running `print(dataset_join_query.statement)`, the `sqlalchemy` code "compiles" into the following:
```sql
SELECT dois_table.doi, library_holdings_data.full_text_indicator 
FROM dois_table
JOIN library_holdings_data 
ON dois_table.database_id = library_holdings_data.doi_foreign_key 
WHERE library_holdings_data.primary_key IN (
    SELECT max(library_holdings_data.primary_key) AS max_1 
    FROM library_holdings_data
    GROUP BY library_holdings_data.doi_foreign_key
)
GROUP BY dois_table.doi
```
But, in this case, the `SORT` is implicit in the fact that the highest `primary_key` for each group will be the newest record for that group. So, we could change this into
```sql
SELECT MAX(dois_table.doi), library_holdings_data.full_text_indicator 
FROM dois_table
JOIN library_holdings_data 
ON dois_table.database_id = library_holdings_data.doi_foreign_key 
GROUP BY dois_table.doi
```
but then there would be no guarantee that the two columns being returned would always be from the same row.

So, we could go back to using a subquery, which I started to write out... and then ended up writing out the same basic approach as in the current `sqlalchemy` code. That seems clearer to me, as I write it out; does it to you, too?

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
dhimmel: > would allow use on non-binary files

All files are binary in the sense that they're just a bunch of bytes. When you open a text file, you're opening a binary file that has an encoding which allows you to translate it to text.

If you run this function with `is_the_file_binary=True`, `shutil.copyfileobj` will throw an error:

```
TypeError: a bytes-like object is required, not 'str'
```

This is because `output_file_location` is always opened in binary mode with the current code. You could open it in text mode, and then Python would default to UTF-8 encoding. But this would potentially re-encode the text file.

So in short, when you use a utility to compress a file (like `gzip` or `xz`) it is always at the file (bytes) level and not at the text level.

So remove this and keep it simple!

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
dhimmel: Ah looks like you're right that there's no straightforward way to do this. See https://stackoverflow.com/a/7745635/4651668. In fact, it's so common that there's a [`greatest-n-per-group`](https://stackoverflow.com/questions/tagged/greatest-n-per-group) tag!

Anyways, go with whatever you think is best... you know SQL better than I!

dhimmel:
> does GitHub hide comments on old code? I'm having trouble finding correspondence that I vaguely remember us

It sometimes minimizes them and it's hard to search.

> I feel fine switching tack and just tracking code here! 

Let's just do code here... it's very close to mergeable. Then the data pull request can be more focused and hence easy sailing.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: Ha, cheers! : P I'd prefer sticking with this way, then. This is my first time using `sqlalchemy`, and it reads here a little bit more unwieldy than just the straight SQL, but it's consistent with the rest of the code.

library_management_system_downloader/copy_and_compress_database_and_extract_tsv.py
jglev: Ah, ok! Thank you for the explanation. Removed in b707e54!

jglev:
I've deleted the example dataset as of f56f01c2.

jglev:
(I also re-ran the code and made sure that it all runs without error -- as expected, it does produce a compressed copy of the database and a compressed TSV of the output data, which have matching numbers of records.)

jglev:
I think at this point, all review comments on this PR have been answered. Does that look correct to you, as well, @dhimmel?

.gitignore
dhimmel: `.gitignore` takes precedent over `.gitattributes`. Probably what happened is you tracked it before adding it to ignore. In the next PR, you will probably have to change this line to remove `*`, so it only ignore the uncompressed.

dhimmel (approved):
Nice! This is a big milestone!

jglev:
Excellent, thank you! I'll get to downloading!

PR #20, opened 2018-01-17
Add results of manual calls of 500 DOIs from on- and off-campus

jglev:
This PR contains the completed 500-DOI sample, from on- and off-campus, with no "invalid" cells.

See [this comment forward](https://github.com/greenelab/library-access/issues/15#issuecomment-358366976) re: DOIs which possibly need to be manually edited for the off-campus column, depending on the outcome of that dicussion.


dhimmel:
How risky / hard would it be to manually port the `penn_access_date` and `open_access_date` from the previous dataset. I think the rows are in the same order, so you may be able to copy and paste? This will make `manual-doi-checks-500.tsv` a bit nicer.

Otherwise, looks good to me once we resolve the login-wall determination.

jglev:
> How risky / hard would it be to manually port the penn_access_date and open_access_date from the previous dataset.

Not hard, and not risky, I think (I'll be careful, and post my code for doing so here using a `JOIN` approach rather than a cut-and-paste). I'll take care of it now.

jglev:
This latest commit looks ok / not dangerous to me; @dhimmel, would you mind looking over the code, too, to confirm that it looks ok to you, as well?

dhimmel:
The changes to `manual-doi-checks-500.tsv` in https://github.com/greenelab/library-access/pull/20/commits/5befc95a6af85e2ac39f896ea9c60fc837052f9d look good.

Can you confirm that login-walled articles are not considered access?

jglev:
Following discussion in #15, I can confirm that I've manually change off-campus access to 0 for the following DOIs, as of afba7a1 (there are two commits because I amended the commit log to include an extra DOI and then `git push --force`d):

- https://doi.org/10.2307/3795274
- https://doi.org/10.2307/3726111
- https://doi.org/10.1039/tf9656100383
- https://doi.org/10.2307/3736092
- https://doi.org/10.2307/317747
- https://doi.org/10.2307/1122632
- https://doi.org/10.1001/jama.1965.03080200059025

dhimmel (approved):

PR #21, opened 2018-01-19
Accuracy analysis on 500 random DOIs

dhimmel:
@publicus can you review?


dhimmel:
@publicus I'm going to merge this now, but please still review and I'll address any issues in a follow up PR.

jglev:
@dhimmel, I just saw this this morning. This all looks excellent to me. I have one *very small* note, but it doesn't affect the code working: In Cell 6: `manual_df.penntext == manual_df.penn_access` can be replaced with `penntext_correct`.

I note that with this sample having been created to match the proportions of PennText saying we have access vs. not, the overall PennText accuracy (cell 6) is in line with your earlier estimate.

I appreciate that you've created the "DOIs with errors" tables -- my plan is that when I have the time this week, for the Library to better understand the issue with our OpenURL resolver, I'll do an analysis of the metadata [that we saved](https://github.com/dhimmel/library-access/blob/be901deb405a63458915caa5061a6136ae4b1b89/data/library_coverage_xml_and_fulltext_indicators.db.xz) for these DOIs, vs. non-problem DOIs (Is it specific publishers? Consistent journals? Is it open-access status? Is it specific date ranges? etc.). That's one way that the database we downloaded *can* be put to immediate use.

jglev:
I do have one question, for my own Python learning: In cell 4, `assert manual_df.query("open_access == 1").penn_access.all()` is a debugging assertion? Which will stop the cell from running if the query doesn't return any rows?

dhimmel:
> a debugging assertion?

Yeah, it's a safety check to ensure that a DOI with open access but not penn access doesn't go unnoticed. Run `assert False` in a python, and you'll see what happens if the statement is not true.

jglev:
Thanks!

Issue #14, opened 2017-12-01
Reminder to add metadata to code files


jglev:
I'm opening this issue as a reminder that, per the [Lab onboarding documentation checklist](http://greenelab-onboarding.readthedocs.io/en/latest/#code-review-checklist), we need to do the following at some point for code files:

1. Sign the files
2. Add a docstring to each of the files
3. For smaller functions, add a docstring

dhimmel:
Regarding "signing the files", we use the git history to take care of that. For example, see [GitHub's blame](https://github.com/greenelab/library-access/blame/172622f8b2a2f04a9284927227da36ce7732e793/library_management_system_downloader/run_api_download_and_parse_results.py) for a file in this repo.

Regarding documentation, we do want to make things clearly documented. Usually, you want to write the doc as you make the files and functions. Sometimes code is clear and doesn't necessarily need docstrings. Are there specific areas of the code where documentation is lacking?

I think the main area where this repo is lacking documentation is in the README. We should update the README to better describe:

1. what the repository does
2. how to use, rerun, or modify it
3. the repository contents (directories and files)

As a reference, you can look at [`greenelab/crossref`](https://github.com/greenelab/crossref) to see how we might want to bolster the README.

dhimmel:
@publicus sort of related to the above discussion, it would be good to add a note to the README letting users know how to rerun this analysis and to modify for another university.

jglev:
@dhimmel Agreed! I can work on that, likely on Friday morning. Does that timing work for you?

dhimmel:
> Agreed! I can work on that, likely on Friday morning. Does that timing work for you?

Sure this is not a time sensitive change although it'd be nice to get it done.

dhimmel:
Closed by https://github.com/greenelab/library-access/pull/26

PR #26, opened 2018-03-01
Adding more descriptive information to readme

jglev:
This follows a [discussion](https://github.com/greenelab/library-access/issues/14#issuecomment-362068564) in #14. A bit late in finishing it up, but I hope it will still be useful for readers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,29 @@ This repository is cataloging University Library access to scholarly literature.
 Scholarly articles are identified using their DOIs.
 The impetus for this project was [this discussion](https://github.com/greenelab/scihub-manuscript/issues/21 "Potential followup: comparison to authorized access") on the Sci-Hub Coverage Study.
 
-## Environment
+## Using the Code
+
+**The code files in this repository assume that your working directory is set to the top-level directory of this repository.**
+
+### Contents of this Repository, and the Order of Their Use
+
+- `LICENSE-*.md`: License text to accompany the [License](#License) section of this Readme below.
+- `environment.yml`: Conda environment file (see [Environment](#environment) below).
+- `.gitattributes`: File with information for tracking files using [Git Large File Storage (LFS)](https://git-lfs.github.com/).
+- `library_management_system_downloader` contains the following scripts, to be used in the following order:
+	- `downloader_configuration_file_TEMPLATE.py` should be copied to `downloader_configuration_file.py` and edited for your own institution's OpenURL resolver (These scripts were specifically tested using the OpenURL resolver that comes with Ex Libris' Alma management software).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,7 +6,31 @@ This repository is cataloging University Library access to scholarly literature.
...
+
+**The code files in this repository assume that your working directory is set to the top-level directory of this repository.**
+
+### Contents of this Repository, and the Order of Their Use
+
+- `LICENSE-*.md`: License text to accompany the [License](#License) section of this Readme below.
+- `environment.yml`: Conda environment file (see [Environment](#environment) below).
+- `.gitattributes`: File with information for tracking files using [Git Large File Storage (LFS)](https://git-lfs.github.com/).
+- `library_management_system_downloader` contains the following scripts, to be used in the following order:
+	- `downloader_configuration_file_TEMPLATE.py` should be copied to `downloader_configuration_file.py` and edited for your own institution's OpenURL resolver (These scripts were specifically tested using the OpenURL resolver that comes with Ex Libris' Alma management software).
```
dhimmel: For ordered things you may want to use a numbered list `1.`?

dhimmel:
Yep. will be useful. Let me know when you're done adding commits on your end.

README.md
jglev: Sure! Changed in b70710.

jglev:
I think that this level of info. is enough to get others going. Do you agree?

dhimmel:
> I think that this level of info. is enough to get others going. Do you agree?

It seems like the most important thing is undocumented -- what is `api_base_url`? How would someone go about getting one for their university?

jglev:
Hmm, I can see how the documentation as it is could be confusing for a reader who's coming in from scratch to parse. This is a documentation difficulty; `api_base_url` is going to be very different from institution to institution, because it'll be based on whomever that institution contracts with for an OpenURL resolver. And finding out what that base URL is will be different for each institution, as it'll involve asking the local development team (e.g., in the library) for help with it / documentation for it. As we've talked about, it's also the case that different OpenURL resolvers may return slightly different formats of data.

Do you think it would be useful to state the above explicitly in the Readme? If so, do you think that that's sufficient for communicating to users?

dhimmel:
> Do you think it would be useful to state the above explicitly in the Readme? If so, do you think that that's sufficient for communicating to users?

Yes.

jglev:
Is 5f7d5a0 in line with what you had in mind, too?

dhimmel (approved):

dhimmel:
Squash merged into https://github.com/greenelab/library-access/commit/310d718f8af0f2e519b84dc7ef72c660b858828e

